PR #6, opened 2019-05-01
Added initial ARM template

dalemorson:
This is the initial ARM template to deploy the following resources:

- App Service Plan
- App Service Certificate
- App Service
- App Insights

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -2,22 +2,28 @@
     "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
     "contentVersion": "1.0.0.0",
     "parameters": {
-        "environmentName": {
+        "environmentFullName": {
             "type": "string",
             "metadata": {
                 "description": "The name of the environment."
+            }
+        },
+        "resourceEnvironmentName": {
+            "type": "string",
+            "metadata": {
+                "description": "The name of the resource environment."
+            }
+        },
+        "configurationStorageConnectionString": {
+            "type": "string",
+            "metadata": {
+                "description": "The connection string to connect to the table configuration storage account."
             }
         },
         "serviceName": {
             "type": "string",
             "metadata": {
                 "description": "The name of the service."
-            }
-        },
-        "sharedResourceGroup": {
-            "type": "string",
-            "metadata": {
-                "description": "The name of the shared resource group."
             }
         },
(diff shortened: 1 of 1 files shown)


azure/template.json
```diff
@@ -0,0 +1,211 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "environmentName": {
```
EwanNoble: This requires two different values as resource naming differs from runtime environment name

azure/template.json
```diff
@@ -0,0 +1,211 @@
...
+                                "value": "[reference('application-insights').outputs.InstrumentationKey.value]"
+                            },
+                            {
+                                "name": "ASPNETCORE_ENVIRONMENT",
+                                "value": "[toUpper(parameters('environmentName'))]"
+                            },
+                            {
+                                "name": "WEBSITE_LOAD_CERTIFICATES",
+                                "value": "[reference('app-service-certificate').outputs.certificateThumbprint.value]"
+                            }
```
EwanNoble: The app settings need the configuration connection string

azure/template.json
```diff
@@ -0,0 +1,211 @@
...
+                    "aspInstances": {
+                        "value": "[parameters('aspInstances')]"
+                    }
+                }
+            }
+        },
+        {
+            "apiVersion": "2017-05-10",
+            "name": "app-service-certificate",
+            "resourceGroup": "[parameters('sharedResourceGroup')]",
```
EwanNoble: Why is this certificate going to the shared resource group?

azure/template.json
dalemorson: I've updated the template to use environmentFullName and resourceEnvironmentName parameters correctly.

azure/template.json
dalemorson: I've added in the configurationStorageConnectionString parameter and added this into the appServiceAppSettings parameters for the App Service.

azure/template.json
dalemorson: Removed the app-service-certificate deployment from being deployed to a shared resource group as there is a new app service plan being deployed specifically for this app service. If the app service were being deployed to a shared app service plan, then I could of used this method.

PR #12, opened 2019-06-25
CV-185 Add add-draft-apprenticeship page

iangledhill:


src/SFA.DAS.EmployerCommitmentsV2/DependencyResolution/DefaultRegistry.cs
```diff
@@ -0,0 +1,20 @@
+﻿using SFA.DAS.ProviderUrlHelper;
```
Pauljgraham: does this belong in the WebRegistry().

src/SFA.DAS.EmployerCommitmentsV2/DependencyResolution/Shared/CommitmentsSharedRegistry.cs
```diff
@@ -0,0 +1,25 @@
...
+                {
+                    scan.AssembliesFromApplicationBaseDirectory(a => a.GetName().Name.StartsWith("SFA.DAS.Commitments.Shared"));
+                    scan.RegisterConcreteTypesAgainstTheFirstInterface();
+                });
+
+            For<ICommitmentsService>().Use<CommitmentsService>().Singleton();
+            IncludeRegistry<ApprenticeshipInfoServiceRegistry>();
+            IncludeRegistry<CommitmentsApiClientRegistry>();
+            IncludeRegistry<EncodingRegistry>();
+        }
```
Pauljgraham: Does this class belong in the CommitmentShared Project? The problem with having it here is we need to know how it's internally built

src/SFA.DAS.EmployerCommitmentsV2/Extensions/IIDentityExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace SFA.DAS.EmployerCommitmentsV2.Web.Extensions
+{
+    public static class IIdentityExtensions
+    {
+        public static string Upn(this IIdentity identity)
+        {
+            switch (identity)
+            {
+                case ClaimsIdentity claimsIdentity:
+                    return claimsIdentity.Claims.FirstOrDefault(claim => claim.Type == EmployeClaims.Upn)?.Value;
```
Pauljgraham: I think this claim type is only applicable for the Provider not the employer (different IDAMS)

src/SFA.DAS.EmployerCommitmentsV2.Web/SFA.DAS.EmployerCommitmentsV2.Web.csproj
```diff
@@ -4,30 +4,61 @@
...
+    <PackageReference Include="WebEssentials.AspNetCore.CdnTagHelpers" Version="1.0.16" />
     <PackageReference Include="WindowsAzure.Storage" Version="9.3.3" />
   </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\SFA.DAS.EmployerCommitmentsV2\SFA.DAS.EmployerCommitmentsV2.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="SFA.DAS.Commitments.Shared">
```
Pauljgraham: I assume this should be removed as we have a nuget reference

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/DraftApprenticeshipController.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            _employerCommitmentsService = employerCommitmentsService;
+            _addDraftApprenticeshipToCohortRequestMapper = addDraftApprenticeshipToCohortRequestMapper;
+            _urlHelper = urlHelper;
+            _trainingProgrammeApiClient = trainingProgrammeApiClient;
+        }
+
+        [HttpGet]
+        [Route("add")]
+        public async Task<IActionResult> AddDraftApprenticeship(ReservationsAddDraftApprenticeshipRequest request)
```
Pauljgraham: ReservationsAddDraftApprenticeshipRequest  doesn't implement the IAuthorization interface so cohortId will not be populated unless it's passed in (which my be ok at the moment)

src/SFA.DAS.EmployerCommitmentsV2.Web/DependencyResolution/IoC.cs
```diff
@@ -8,8 +10,10 @@ public static class IoC
     {
         public static void Initialize(Registry registry)
         {
-            registry.IncludeRegistry<CommitmentsApiClientRegistry>();
-            registry.IncludeRegistry<ConfigurationRegistry>();
+            registry.IncludeRegistry<DefaultRegistry>();
```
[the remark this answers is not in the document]
iangledhill: The precedence of the two registries is different; the default happens first, so can be overridden/replaced by any of the subsequent registries. The web registry (which is empty, so is not actually needed) overrides/replaces any of the preceding registries. I wanted mine to run first, so it is in the appropriate place.

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/DraftApprenticeshipController.cs
```diff
@@ -0,0 +1,109 @@
...
+using SFA.DAS.EmployerCommitmentsV2.Extensions;
+using SFA.DAS.EmployerCommitmentsV2.Web.Models;
+using SFA.DAS.EmployerCommitmentsV2.Web.Requests;
+using SFA.DAS.ProviderUrlHelper;
+
+namespace SFA.DAS.ProviderCommitments.Web.Controllers
+{
+    
+    [Route("accounts/{hashedAccountId}/apprentices/")]
+    public class DraftApprenticeshipController : Controller
```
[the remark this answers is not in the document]
iangledhill: Have added some tests

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/DraftApprenticeshipController.cs
```diff
@@ -0,0 +1,109 @@
...
+                return View(model);
+            }
+        }
+
+        private async Task AddLegalEntityAndCoursesToModel(DraftApprenticeshipViewModel model)
+        {
+            var cohortDetail = await _employerCommitmentsService.GetCohortDetail(model.CohortId.Value);
+            var courses = await GetCourses(cohortDetail);
+
+            model.Employer = cohortDetail.LegalEntityName;
```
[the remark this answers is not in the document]
iangledhill: Removed

src/SFA.DAS.EmployerCommitmentsV2/DependencyResolution/Shared/CommitmentsSharedRegistry.cs
iangledhill: Ah, it is in the shared (and being used from there). This class isn't used - have removed

src/SFA.DAS.EmployerCommitmentsV2.Web/SFA.DAS.EmployerCommitmentsV2.Web.csproj
iangledhill: Removed

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/DraftApprenticeshipController.cs
iangledhill: Believe you meant IAuthorizationContextModel - have added. No idea what this interface does - it has no methods, which presumably means it's a marker interface. I wonder what Uncle Bob would say.

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/DraftApprenticeshipController.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            try
+            {
+                await _employerCommitmentsService.AddDraftApprenticeshipToCohort(model.CohortId.Value, request);
+                var cohortDetailsUrl = $"{model.ProviderId}/apprentices/{model.CohortReference}/Details";
+                return Redirect("add");
+            }
+            catch (CommitmentsApiModelException ex)
+            {
+                ModelState.AddModelExceptionErrors(ex);
```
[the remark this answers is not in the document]
iangledhill: They could've been put there previously, but weren't so presumably whoever wrote it thought that wasn't the appropriate place to put it.

src/SFA.DAS.EmployerCommitmentsV2/DependencyResolution/DefaultRegistry.cs
iangledhill: I think this was already raised?

src/SFA.DAS.EmployerCommitmentsV2/Extensions/IIDentityExtensions.cs
iangledhill: I copied this across pre-emptively for CV-190. I'll check this as part of that story (I'm working on it now) and fix it then if the claims are different.

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/DraftApprenticeshipController.cs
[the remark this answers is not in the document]
chrisfoster76: Seems like they are in the right place already

src/SFA.DAS.EmployerCommitmentsV2/Extensions/IIDentityExtensions.cs
chrisfoster76: The name space is wrong for this class, and the filename is ever-so-slightly-wrong, as is the `EmployeClaims` class.

The claims themselves (looking at v1 code) seem to be:
`  public static class DasClaimTypes
  {
    private const string BaseUrl = "http://das/employer/identity/claims/";
    public const string Id = "http://das/employer/identity/claims/id";
    public const string Email = "http://das/employer/identity/claims/email_address";
    public const string GivenName = "http://das/employer/identity/claims/given_name";
    public const string FamilyName = "http://das/employer/identity/claims/family_name";
    public const string DisplayName = "http://das/employer/identity/claims/display_name";
    public const string RequiresVerification = "http://das/employer/identity/claims/requires_verification";
  }`
plus, "sub" which appears to be the User Id.

PR #18, opened 2019-07-16
Cv 318 edit page redirects to v1 view page if cohort with provider

Pauljgraham:


Pauljgraham:
This is being merged via the CV-247 edit draft apprenticeship branch

PR #24, opened 2019-07-23
CV-291 Added delete button to edit draft apprenticeship page

elliotritchie:


src/SFA.DAS.EmployerCommitmentsV2.Web/Views/DraftApprenticeship/EditDraftApprenticeship.cshtml
```diff
@@ -16,10 +17,10 @@
             @Html.AntiForgeryToken()
             <partial name="DraftApprenticeship" for="@Model" />
             <button type="submit" class="govuk-button">Save</button>
-            @*<a class="govuk-button govuk-button--warning das-float-right" href="@Url.ProviderApprenticeshipServiceLink($"{Model.ProviderId}/apprentices/{Model.CohortReference}/{Model.DraftApprenticeshipHashedId}/delete")">Delete</a>
-        <p class="govuk-body  govuk-!-margin-bottom-0">
-            <a class="govuk-body govuk-link" href="@Url.ProviderApprenticeshipServiceLink($"{Model.ProviderId}/apprentices/{Model.CohortReference}/details")">Cancel</a>
-        </p>*@
+            <a class="govuk-button govuk-button--warning das-float-right" href="@LinkGenerator.CommitmentsLink($"commitments/accounts/{Model.AccountHashedId}/apprentices/{Model.CohortReference}/apprenticeships/{Model.DraftApprenticeshipHashedId}/delete")">Delete</a>
```
chrisfoster76: I don't think that the link should start with "commitments/" - I believe this is already there in the config. I am not 100% sure however. But, contrast with CreateCohortWithDraftApprenticeshipController.cs, line 84:

```var reviewYourCohort = _linkGenerator.CommitmentsLink($"accounts/{model.AccountHashedId}/apprentices/{newCohort.CohortReference}/details");```

chrisfoster76: Okay, just checked, and the /commitments part of the route is always there in the config, so it's part of the baseurl and should not be included.

PR #26, opened 2019-07-25
CV-312 confirm traning provider details

dashton82:


src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Shared/_ConfirmProviderHiddenPartial.cshtml
```diff
@@ -0,0 +1,11 @@
...
+
+
+<input type="hidden" id="@nameof(ConfirmProviderViewModel.AccountHashedId)" name="@nameof(ConfirmProviderViewModel.AccountHashedId)" value="@Model.AccountHashedId"/>
+<input type="hidden" id="@nameof(ConfirmProviderViewModel.EmployerAccountLegalEntityPublicHashedId)" name="@nameof(ConfirmProviderViewModel.EmployerAccountLegalEntityPublicHashedId)" value="@Model.EmployerAccountLegalEntityPublicHashedId"/>
+<input type="hidden" id="@nameof(ConfirmProviderViewModel.ReservationId)" name="@nameof(ConfirmProviderViewModel.ReservationId)" value="@Model.ReservationId"/>
+<input type="hidden" id="@nameof(ConfirmProviderViewModel.CourseCode)" name="@nameof(ConfirmProviderViewModel.CourseCode)" value="@Model.CourseCode"/>
+<input type="hidden" id="@nameof(ConfirmProviderViewModel.StartMonthYear)" name="@nameof(ConfirmProviderViewModel.StartMonthYear)" value="@Model.StartMonthYear"/>
+<input type="hidden" id="@nameof(ConfirmProviderViewModel.ProviderId)" name="@nameof(ConfirmProviderViewModel.ProviderId)" value="@Model.ProviderId"/>
+<input type="hidden" id="@nameof(ConfirmProviderViewModel.ProviderName)" name="@nameof(ConfirmProviderViewModel.ProviderName)" value="@Model.ProviderName"/>
```
Pauljgraham: For these inputs is it possible to user the helper Html.HiddenFor(....) for each element.

I think you;re doing this anyway, but EmployerAccountLegalEntityPublicHashedId should be AccountLegalEntityHashedId

ProviderName can be added to the view model or retrieved each time (both ways are valid solutions), I think we need to agree which technique we should standardise on

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/CreateCohort/ConfirmProvider.cshtml
```diff
@@ -1,7 +1,57 @@
...
+                    </legend>
+                    @if (!ViewData.ModelState.IsValid && ViewData.ModelState.ContainsKey(nameof(ConfirmProviderViewModel.UseThisProvider)))
+                    {
+                        <span id="error-message-@nameof(ConfirmProviderViewModel.UseThisProvider)" class="govuk-error-message">
+                            <span class="govuk-visually-hidden">Error:</span> @ViewData.ModelState[nameof(ConfirmProviderViewModel.UseThisProvider)].Errors.First().ErrorMessage
+                        </span>
+                    }
+                    <div class="govuk-radios">
+                        <div class="govuk-radios__item">
+                            <input class="govuk-radios__input" id="UseThisProvider" name="UseThisProvider" type="radio" value="true">
```
Pauljgraham: I don't know but is there a helper for radio buttons? Like the other controls

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/CreateCohort/ConfirmProvider.cshtml
```diff
@@ -1,7 +1,57 @@
...
+        <p class="govuk-body-l">UK Provider Reference Number <strong>'@Model.ProviderId'</strong> matches:</p>
+        
+        <h2 class="govuk-heading-l govuk-!-margin-bottom-0">@Model.ProviderName</h2>
+        <dl class="das-definition-list das-definition-list--inline  das-definition-list--muted govuk-!-margin-bottom-7">
+            <dt class="das-definition-list__title">UKPRN</dt>
+            <dd class="das-definition-list__definition">@Model.ProviderId</dd>
+        </dl>
+
+        <form method="post" asp-controller="CreateCohort" asp-action="ConfirmProvider">
+            <partial name="_ConfirmProviderHiddenPartial" model="@Model" />
```
Pauljgraham: If this is only used on this page. Should it be in the the shared area

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/CreateCohort/ConfirmProvider.cshtml
```diff
@@ -1,7 +1,57 @@
-﻿
+﻿@using SFA.DAS.EmployerCommitmentsV2.Web.Models.CreateCohort
+@model SFA.DAS.EmployerCommitmentsV2.Web.Models.CreateCohort.ConfirmProviderViewModel
```
Pauljgraham: could we drop the namespace on the model? As it's in the 'using' line

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/CreateCohort/ConfirmProviderViewModel.cs
```diff
@@ -1,10 +1,15 @@
...
+using System.ComponentModel.DataAnnotations;
 
 namespace SFA.DAS.EmployerCommitmentsV2.Web.Models.CreateCohort
 {
     public class ConfirmProviderViewModel : IndexViewModel
     {
         public long ProviderId { get; set; }
+        public string ProviderName { get; set; }
+
+        [Required(ErrorMessage = "Select a training provider")]
```
Pauljgraham: Is this needed if we are using FluentValidation?

src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/SelectProviderConfirmProviderMapper.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public SelectProviderViewModel Map(ConfirmProviderViewModel source)
+        {
+            return new SelectProviderViewModel
+            {
+                ProviderId = source.ProviderId.ToString(),
+                ReservationId = source.ReservationId,
+                CourseCode = source.CourseCode,
+                AccountHashedId = source.AccountHashedId,
+                EmployerAccountLegalEntityPublicHashedId = source.EmployerAccountLegalEntityPublicHashedId,
```
Pauljgraham: this should be AccountLegalEntityHashedId

src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/ConfirmProviderViewModelMapper.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace SFA.DAS.EmployerCommitmentsV2.Web.Mappers
+{
+    public class ConfirmProviderViewModelMapper : IMapper<ConfirmProviderRequest, ConfirmProviderViewModel>
+    {
+        public ConfirmProviderViewModel Map(ConfirmProviderRequest source)
+        {
+            return new ConfirmProviderViewModel
+            {
+                AccountHashedId = source.AccountHashedId,
+                EmployerAccountLegalEntityPublicHashedId = source.EmployerAccountLegalEntityPublicHashedId,
```
Pauljgraham: this should be AccountLegalEntityHashedId

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CreateCohortController.cs
```diff
@@ -88,21 +98,57 @@ public async Task<IActionResult> SelectProvider(SelectProviderViewModel request)
...
+
+            return View(model);
+        }
+
+        [Route("confirm-provider")]
+        [HttpPost]
+        public IActionResult ConfirmProvider(ConfirmProviderViewModel request)
         {
-            return View(request);
+            var validationResult = _confirmProviderViewModelValidator.Validate(request);
```
Pauljgraham: I don't think this needs to called explicitly, it should occur automatically in the pipeline

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CreateCohortController.cs
```diff
@@ -88,21 +98,57 @@ public async Task<IActionResult> SelectProvider(SelectProviderViewModel request)
...
-                _logger.LogError($"Failed '{nameof(CreateCohortController)}-{nameof(SelectProvider)}': {nameof(ex.StatusCode)}='{ex.StatusCode}', {nameof(ex.ReasonPhrase)}='{ex.ReasonPhrase}'");
+
+                _logger.LogError(
+                    $"Failed '{nameof(CreateCohortController)}-{nameof(SelectProvider)}': {nameof(ex.StatusCode)}='{ex.StatusCode}', {nameof(ex.ReasonPhrase)}='{ex.ReasonPhrase}'");
             }
             catch (Exception ex)
             {
-                _logger.LogError($"Failed '{nameof(CreateCohortController)}-{nameof(SelectProvider)}': {nameof(ex.Message)}='{ex.Message}', {nameof(ex.StackTrace)}='{ex.StackTrace}'");
+                _logger.LogError(
+                    $"Failed '{nameof(CreateCohortController)}-{nameof(SelectProvider)}': {nameof(ex.Message)}='{ex.Message}', {nameof(ex.StackTrace)}='{ex.StackTrace}'");
```
Pauljgraham: Should we re-throw an un-handled exception?

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CreateCohortController.cs
```diff
@@ -30,15 +34,21 @@ public class CreateCohortController : Controller
             IMapper<IndexRequest, IndexViewModel> indexViewModelMapper,
             IMapper<SelectProviderRequest, SelectProviderViewModel> selectProviderViewModelMapper,
             IMapper<SelectProviderViewModel, ConfirmProviderRequest> confirmProviderRequestMapper,
+            IMapper<ConfirmProviderRequest, ConfirmProviderViewModel> confirmProviderViewModelMapper,
+            IMapper<ConfirmProviderViewModel, SelectProviderViewModel> selectProviderFromConfirmMapper,
             IValidator<SelectProviderViewModel> selectProviderViewModelValidator,
+            IValidator<ConfirmProviderViewModel> confirmProviderViewModelValidator,
```
Pauljgraham: We should be able to remove the Validators

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CreateCohortController.cs
```diff
@@ -21,7 +22,10 @@ public class CreateCohortController : Controller
         private readonly IMapper<IndexRequest, IndexViewModel> _indexViewModelMapper;
         private readonly IMapper<SelectProviderRequest, SelectProviderViewModel> _selectProviderViewModelMapper;
         private readonly IMapper<SelectProviderViewModel, ConfirmProviderRequest> _confirmProviderRequestMapper;
+        private readonly IMapper<ConfirmProviderRequest, ConfirmProviderViewModel> _confirmProviderViewModelMapper;
+        private readonly IMapper<ConfirmProviderViewModel, SelectProviderViewModel> _selectProviderFromConfirmMapper;
         private readonly IValidator<SelectProviderViewModel> _selectProviderViewModelValidator;
+        private readonly IValidator<ConfirmProviderViewModel> _confirmProviderViewModelValidator;
```
Pauljgraham: We shouldn't need the validators

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/CreateCohort/ConfirmProvider.cshtml
dashton82: done.

src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/ConfirmProviderViewModelMapper.cs
dashton82: This is fixed in #29

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CreateCohortController.cs
dashton82: Not sure this needs re-throwing - its handled and logged, it will fall through to the `RedirectToAction("Error","Error")` at the bottom of the action

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CreateCohortController.cs
dashton82: done

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CreateCohortController.cs
dashton82: done

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CreateCohortController.cs
dashton82: done

src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/SelectProviderConfirmProviderMapper.cs
dashton82: Fixed in #29

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/CreateCohort/ConfirmProvider.cshtml
dashton82: Its following what the previous pages have done, i think its been done like this to allow us to consolidate this into one partial later on

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Shared/_ConfirmProviderHiddenPartial.cshtml
dashton82: I'm happy to change, however I think we should be consistent in what we are doing on each page, and having a mismatch of razor and tag helper code will only further obfuscate the intention from a front end dev who may not know. We aren't using the Html.BeginForm or a lot of the other ones, so if we are to change this one then we should probably make it uniform for all elements on all views.

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/CreateCohort/ConfirmProvider.cshtml
dashton82: Same as above comment

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/CreateCohort/ConfirmProviderViewModel.cs
dashton82: Removed annotation

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Shared/_ConfirmProviderHiddenPartial.cshtml
dashton82: As agreed with Chris these have been changed to 
```
<input type="hidden" asp-for="FieldName">
```

PR #29, opened 2019-07-29
CV-193 create cohort assign

dashton82:


src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CreateCohortController.cs
```diff
@@ -131,6 +143,7 @@ public async Task<IActionResult> ConfirmProvider(ConfirmProviderRequest request)
 
         [Route("confirm-provider")]
         [HttpPost]
+        [ValidateAntiForgeryToken]
         public IActionResult ConfirmProvider(ConfirmProviderViewModel request)
         {
             var validationResult = _confirmProviderViewModelValidator.Validate(request);
```
chrisfoster76: This explicit call can be removed

dashton82: Done, removed all and added global filter

PR #53, opened 2019-09-18
CV-9 Allow Employer to add another apprentice

Pauljgraham:


src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Mappers/Cohort/DetailsViewModelMapperTests.cs
```diff
@@ -6,93 +6,178 @@
...
         }
 
         [Test]
-        public void MessageIsMappedCorrectly()
+        public async Task TransferSenderHashedIdIsEncodedCorrectlyWhenThereIsAValue()
         {
-            Assert.AreEqual(_cohort.LatestMessageCreatedByProvider, _result.Message);
+            var fixture = new DetailsViewModelMapperTestsFixture().SetTransferSenderId(123);
+            var result = await fixture.Map();
+            Assert.AreEqual("X123X", result.TransferSenderHashedId);
```
iangledhill: Just a minor observation - it's not obvious where the expected value comes from - a quick inspection soon reveals the setup on the encoding service is coded to do this. I wonder whether it might be a little clearer if the expected value was passed into the SetTransferSenderId as a second parameter? The setup on the encoding service could be done there, which would also restrict the encoder to the expected values (rather than accepting any long as it currently does).

Pauljgraham: good idea. will do

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Cohort/Details.cshtml
```diff
@@ -57,71 +58,76 @@
...
+        <tbody class="govuk-table__body">
+            <tr class="govuk-table__row">
+                <th class="govuk-table__header das-table-cell-width-80" scope="row">Total apprenticeship training price <br>(excluding VAT)</th>
+                <td class="govuk-table__cell govuk-table__header--numeric das-table-cell-width-10"><strong>&pound;42,000</strong></td>
+                <td class="govuk-table__cell das-table-cell-width-10" aria-hidden="true"></td>
+            </tr>
+        </tbody>
+    </table>*@
+
+@if (Model.CanAmendCohort)
```
chrisfoster76: @Pauljgraham what is the reason for this condition? In the current epic, we are building an editable version of this page, so I don't think we need it. Also, the logic seems to use the old EditStatus field. We should not be exposing this field through our v2 api.

PR #56, opened 2019-09-19
CV-433 Standardise mapper names on <Target>Mapper

iangledhill:
This is a technical task that was picked up during CV-385 to normalize the names of the mappers. The normalized name is <TargetName>Mapper. Where there are multiple sources that all map to the same target then the mapper class will implement an IMapper<from, to> for each source/target combination.

*Replaces CV-385 PR*


chrisfoster76:
Merged into CV-380

PR #57, opened 2019-09-19
Cv 20 add grouping

iangledhill:


src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/Cohort/DetailsViewModelMapper.cs
```diff
@@ -3,28 +3,32 @@
...
             var cohortTask = _commitmentsApiClient.GetCohort(source.CohortId);
             var draftApprenticeshipsTask = _commitmentsApiClient.GetDraftApprenticeships(source.CohortId);
 
-            await Task.WhenAll(new List<Task> { cohortTask, draftApprenticeshipsTask });
+            await Task.WhenAll(cohortTask, draftApprenticeshipsTask);
 
-            var cohort = await cohortTask;
-            var draftApprenticeships = (await draftApprenticeshipsTask).DraftApprenticeships;
+            var cohort = cohortTask.Result;
+            var draftApprenticeships = draftApprenticeshipsTask.Result.DraftApprenticeships;
```
Pauljgraham: I read an article which said whilst it was perfectly ok to call .Result in the above circumstance the await style was more appropriate and did exactly the same thing (with no additional cost), as the await will just get the result.

The only other benefit is from a personal POV is I usually look twice when I see a .Result whilst the await looks slight more 'normal'

PR #58, opened 2019-09-19
Cv 403 delete this group

Pauljgraham:
Still waiting for James King to give us the layout for the LHS/RHS add and delete links


src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Mappers/DraftApprenticeship/AddDraftApprenticeshipViewModelMapperTests.cs
```diff
@@ -101,11 +101,11 @@ public void ProviderNameIsMappedCorrectly()
             Assert.AreEqual(_cohort.ProviderName, _result.ProviderName);
         }
 
-        [TestCase(true)]
-        [TestCase(false)]
-        public async Task CoursesAreMappedCorrectly(bool fundedByTransfer)
+        [TestCase(123, true)]
+        [TestCase(null, false)]
```
iangledhill: I wonder if adding zero as a test case for transfer sender id would make it clear whether zeros are treated as funded by transfer, to support UT's aim of documenting expected bahaviour?

PR #59, opened 2019-09-20
CV-428 Make page title plural aware

iangledhill:


src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Mappers/Cohort/DetailsViewModelMapperTests.cs
```diff
@@ -155,6 +164,27 @@ public void DraftApprenticeshipsAreMappedCorrectly()
...
+            SetDraftApprenticeships(draftApprenticeships);
+
+            // Act_
+            var result = await _mapper.Map(_source);
+
+            // Assert
+
+            Assert.AreEqual(expectedPageTitle, result.PageTitle);
+        }
+
```
Pauljgraham: I believe I refactored these unit tests in CV-9 to use the Fixture class design we agreed to follow. Would that style be a better way to write this tests as the mapper is getting quite complicated. I think these different approaches will cause a merge problem.

iangledhill: Cheers - I'll fix these tests up when I merge them up.

PR #60, opened 2019-09-24
Cv 404 funding band warnings

Pauljgraham:


src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/Cohort/DetailsViewModelMapper.cs
```diff
@@ -38,37 +42,85 @@ public async Task<DetailsViewModel> Map(DetailsRequest source)
...
                             Cost = a.Cost,
+                            FundingBandCap = fc.FundingBandCap,
                             DateOfBirth = a.DateOfBirth,
                             EndDate = a.EndDate,
                             StartDate = a.StartDate
                         })
-                        .ToList()
-                })
-                .OrderBy(c => c.CourseName)
+                        .Select(a => a)
```
iangledhill: Does the select do anything?

src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/Cohort/DetailsViewModelMapper.cs
```diff
@@ -38,37 +42,85 @@ public async Task<DetailsViewModel> Map(DetailsRequest source)
...
+                Courses = await GroupCourses(draftApprenticeships.ToList())
             };
         }
 
-        private IReadOnlyCollection<DetailsViewCourseGroupingModel> GroupCourses(IEnumerable<DraftApprenticeshipDto> draftApprenticeships)
+        private async Task<IReadOnlyCollection<DetailsViewCourseGroupingModel>> GroupCourses(IList<DraftApprenticeshipDto> draftApprenticeships)
         {
-            return draftApprenticeships
-                .GroupBy(a => new {a.CourseCode, a.CourseName})
+            var apprenticeFundingBandCaps = await Task.WhenAll(draftApprenticeships.Select(async a => new { a.Id, FundingBandCap = await GetFundingBandCap(a.CourseCode, a.StartDate) }));
```
iangledhill: This operation is going to be CPU bound (not I/O) so is not going to benefit from running async. Since this is going to call the funding csap method for all the draft apprenticeships anyway would the code be simpler if the call to get GetFundingBandCap was moved to the L70 and the join replaced with a select?

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Cohort/FundingBandExcessModel.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        private readonly int?[] _fundingBandCapsExceeded;
+
+        public FundingBandExcessModel(int apprenticesExceedingBand, IEnumerable<int?> fundingBandCapsExceeded)
+        {
+            _fundingBandCapsExceeded = fundingBandCapsExceeded.ToArray();
+            ApprenticesExceedingBand = apprenticesExceedingBand;
+        }
+
+        public int ApprenticesExceedingBand { get; }
```
iangledhill: It wasn't immediately obvious to me that this was a count - I did figure this out by looking at the constructor calls but wonder if a better name could be found?

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Cohort/Details.cshtml
```diff
@@ -31,30 +32,61 @@
 {
     <h1>@course.Count x @course.DisplayCourseName</h1>
 
+    if (@course.FundingBandExcess != null)
+    {
+    <div class="govuk-inset-text das-inset-text--highlight">
+        <h3 class="govuk-heading-s">@course.FundingBandExcess.ApprenticesExceedingBand apprenticeships above the funding band maximum</h3>
+        <p>
+            The price for this apprenticeship is above it's
+            @if (course.FundingBandExcess.SingleFundingBandCapExceeded == null)
```
iangledhill: If this logic were in the view model (i.e. add a DisplayFundingBandeExceededMessage) then it could be unit tested.

src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/Cohort/DetailsViewModelMapper.cs
Pauljgraham: unfortunately because GetFundingCap uses an async call, it was easier to execute it this way, otherwise the code became complex.

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Cohort/Details.cshtml
Pauljgraham: This is here for display reasons only, I didn't want to introduce hyper link into a view model

PR #66, opened 2019-10-17
Cv 450 do not allow employers to approve cohorts with missing mandatory fields

PeterDavidWilliams:


src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Cohort/Details.cshtml
```diff
@@ -22,6 +22,23 @@
         </div>
     </div>  
 }
+else
+{
+    if (!Model.EmployerCanApprove && Model.DraftApprenticeshipsCount > 0)
```
chrisfoster76: As per comment on the other PR, a single IsCompleteForEmployer that includes cohorts not being empty will simplify this line

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Cohort/Details.cshtml
```diff
@@ -136,31 +153,41 @@
                         </label>
                     </div>
                 }
-
-                @if (Model.ShowApprovalOption)
+                
+                @if (Model.EmployerCanApprove)
```
chrisfoster76: > Model.EmployerCanApprove _and_ Model.ShowApprovalOption

Can we make this more clear?

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Cohort/Details.cshtml
```diff
@@ -136,31 +153,41 @@
...
+                                     Leave @Model.ProviderName a message (optional)
+                                 </label>
+                                 <textarea class="govuk-textarea" id="approve-details" name="approvemessage" rows="5"></textarea>
+                             </div>
+                         </div>
+                     }
                 }
-                @if (!Model.ShowApprovalOptionMessage)
+                
+                @{ var requestChangesLabel = new string("No, request changes from training provider");  }
```
chrisfoster76: We can probably do away with the variable altogether, and definitely the "new string" bit at least.

PeterDavidWilliams: Agree with the "new string".

This was the best way  I could think of to change the content of the label based on a conditional, suggestions are welcome.

PR #67, opened 2019-10-21
CV-452

hkf-tech:
Replace the read-only 'View your cohort' pages.

### DIFF
--- a/src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Controllers/CohortControllerTests/WhenGettingDetails.cs
+++ b/src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Controllers/CohortControllerTests/WhenGettingDetails.cs
@@ -38,7 +38,7 @@ namespace SFA.DAS.EmployerCommitmentsV2.Web.UnitTests.Controllers.CohortControll
         {
             _fixture.WithParty(withParty);
             await _fixture.GetDetails();
-            _fixture.VerifyResultIsRedirectToV1();
+            _fixture.VerifyResultIsNotRedirectToV1();
         }
 
         public class WhenGettingDetailsTestFixture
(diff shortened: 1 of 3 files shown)


src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Cohort/Details.cshtml
```diff
@@ -5,6 +5,7 @@
 
 @{
     ViewData["Title"] = Model.PageTitle;
+    var isReadOnly = Model.WithParty != SFA.DAS.CommitmentsV2.Types.Party.Employer;
```
chrisfoster76: Can we add this as a property the view model, so that we can avoid code in the razor and cover it with unit tests?

src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Controllers/CohortControllerTests/WhenGettingDetails.cs
```diff
@@ -99,12 +99,11 @@ public void VerifyViewModelIsMappedFromRequest()
                 Assert.AreEqual(expectedTotalCost, _viewModel.TotalCost, "The total cost stored in the model is incorrect");
             }
 
-            public void VerifyResultIsRedirectToV1()
+            public void VerifyResultIsNotRedirectToV1()
```
chrisfoster76: As we're removing the redirect altogether, I suggest we delete this test completely, so that we're not left with tests around things we're _not_ doing.

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Cohort/Details.cshtml
```diff
@@ -93,6 +97,8 @@
     </tbody>
 </table>
 
+@if (!isReadOnly)
```
chrisfoster76: As above, into the view model with this.

PR #76, opened 2019-11-19
Cv 473 Rebuild Draft Cohorts Page V2

ChughAkshay:


src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Cohort/Draft.cshtml
```diff
@@ -0,0 +1,50 @@
...
+                        @foreach (var cohort in Model.Cohorts)
+                        {
+                            var cohortReference = EncodingService.Encode(cohort.CohortId, EncodingType.CohortReference);
+
+                            <tr class="govuk-table__row">
+                                <td class="govuk-table__cell" data-label="Training provider">@cohort.ProviderName</td>
+                                <td class="govuk-table__cell" data-label="Reference">@cohortReference</td>
+                                <td class="govuk-table__cell govuk-table__cell--numeric" data-label="Number of apprentices">@cohort.NumberOfDraftApprentices</td>
+                                <td class="govuk-table__cell govuk-table__cell--numeric" data-label="">
+                                    @Html.ActionLink("Details", "Details", "Cohort", new { @Model.AccountHashedId, cohortReference }, new { @class = "govuk-link", id = "return-link" })
```
chrisfoster76: We should probably not use an Id property repeatedly, and this seems a strange Id to use in any case. We could instead add a css class "details-link" or similar as useful for automation tests.

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Cohort/Draft.cshtml
```diff
@@ -0,0 +1,50 @@
...
+                            <th scope="col" class="govuk-table__header das-table-cell-width-25">Reference</th>
+                            <th scope="col" class="govuk-table__header das-table-cell-width-25 govuk-table__header--numeric">Number of apprentices</th>
+                            <th scope="col" class="govuk-table__header govuk-table__header--numeric das-table-cell-width-25"></th>
+                        </tr>
+                    </thead>
+                    <tbody class="govuk-table__body">
+
+                        @foreach (var cohort in Model.Cohorts)
+                        {
+                            var cohortReference = EncodingService.Encode(cohort.CohortId, EncodingType.CohortReference);
```
chrisfoster76: We could encode the Id in the view model mapper just as easily, and reduce clutter

src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/Cohort/DraftRequestMapper.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            _commitmentsApiClient = commitmentsApiClient;
+            _linkGenerator = linkGenerator;
+        }
+
+        public async Task<DraftViewModel> Map(DraftRequest source)
+        {
+            var request = new GetCohortsRequest {AccountId = source.AccountId};
+            var apiResponse = await _commitmentsApiClient.GetCohorts(request);
+            var cohorts = apiResponse.Cohorts.Where(x => x.WithParty == Party.Employer && x.IsDraft).OrderByDescending(x => x.CreatedOn);
```
chrisfoster76: Can we take this Where expression and replace it with a call to a "GetStatus()" extension method that returns a (new) CohortStatus enum - this would allow all Status calculation to reside in one place.

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Cohort/DraftViewModel.cs
```diff
@@ -0,0 +1,14 @@
...
+using SFA.DAS.CommitmentsV2.Types;
+
+namespace SFA.DAS.EmployerCommitmentsV2.Web.Models.Cohort
+{
+    public class DraftViewModel
+    {
+        public string AccountHashedId { get; set; }
+        public long AccountId { get; set; }
+        public string BackLink { get; set; }
+        public IEnumerable<CohortSummary> Cohorts { get; set; }
```
chrisfoster76: We should remove the reference to the api type here and map to a local one

PR #97, opened 2020-02-07
MF-430 empty state manage apprentices page

scottywakefield:
Includes feature toggle


src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class ApprenticeController : Controller
+    {
+        [Route("", Name = RouteNames.ManageApprentices)]
+        [DasAuthorize(EmployerFeature.ManageApprenticesV2)]
+        public IActionResult Index(string accountHashedId)
+        {
+            if (!ModelState.IsValid)
+            {
+                return BadRequest(ModelState);
+            }
```
dashton82: Dont think we need this

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace SFA.DAS.EmployerCommitmentsV2.Web.Controllers
+{
+    [Route("{accountHashedId}/apprentices")]
+    [SetNavigationSection(NavigationSection.ApprenticesHome)]
+    [DasAuthorize(EmployerUserRole.OwnerOrTransactor)]
+    public class ApprenticeController : Controller
+    {
+        [Route("", Name = RouteNames.ManageApprentices)]
+        [DasAuthorize(EmployerFeature.ManageApprenticesV2)]
+        public IActionResult Index(string accountHashedId)
```
dashton82: Following the pattern, i think this will need to be IndexRequest with accountHashedId within it

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
scottywakefield: removed in MF-431

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
scottywakefield: done in MF-431

PR #100, opened 2020-02-14
MF-431 manage apprentices table

scottywakefield:


src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/IndexRequest.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using SFA.DAS.CommitmentsV2.Types;
+
+namespace SFA.DAS.EmployerCommitmentsV2.Web.Models.Apprentice
+{
+    public class IndexRequest
+    {
+        public string HashedAccountId { get; set; }
+        public int PageNumber { get; set; }
```
PeterDavidWilliams: May not be the place for it, but if we don't set default PageNumber = 1, on initially loading the page the 12 month download restriction is going to apply to the data that is shown

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticesFilterModel.cs
```diff
@@ -0,0 +1,173 @@
...
+        public HtmlString FiltersUsedMessage => this.GetFiltersUsedMessage();
+
+        public int TotalNumberOfApprenticeships { get; set; }
+        public int TotalNumberOfApprenticeshipsFound { get; set; }
+        public int TotalNumberOfApprenticeshipsWithAlertsFound { get; set; }
+        
+        public IEnumerable<PageLink> PageLinks {
+            get
+            {
+                var links = new List<PageLink>();
```
robinoconnor: This seems like a lot of logic to be placed into a Get statement. Would this be better to either extract out into a helper and use that or break it down internally into private methods?

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticesFilterModel.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            return routeData;
+        }
+
+        public Dictionary<string, string> BuildSortRouteData(string sortField)
+        {
+            var routeData = BuildRouteData();
+
+            var reverseSort = !string.IsNullOrEmpty(SortField) 
+                              && SortField.ToLower() == sortField.ToLower()
```
robinoconnor: It might be beneficial to name the parameter sortField into something different to make it easier to distinguish between the Property. It would just make it slightly easier to read code wise

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/DownloadViewModel.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace SFA.DAS.EmployerCommitmentsV2.Web.Models.Apprentice
+{
+    public class DownloadViewModel
+    {
+        public byte[] Content { get; set; }
+        public string Name { get; set; }
+        public string ContentType => "text/csv";
```
robinoconnor: Is Content type ever likely to be anything else as it could be a const if not

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/IndexViewModel.cs
```diff
@@ -0,0 +1,29 @@
...
+            {
+                SortedByHeaderClassName += " das-table__sort--desc";
+            }
+            else
+            {
+                SortedByHeaderClassName += " das-table__sort--asc";
+            }
+        }
+
+        public ApprenticesFilterModel FilterModel { get; set; }
```
robinoconnor: These should be with the other properties of the class unless there is a reason for them alone being after the method?

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticesFilterModel.cs
scottywakefield: I agree with you. Spent some time investigating and my conclusion is both of those options lead to other problems. helper needs injection somewhere - either into the model or the mapper, both look like a misuse of DI. putting into a method just moves the problem somewhere else. it's worth noting that the `PageLinks` property shares logic with `BuildSortRouteData()` function.

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/IndexRequest.cs
scottywakefield: yes, it was there before ... wonder why it's removed.

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticeshipDetailsCsvModel.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        [Name("Apprentice Name")]
+        public string ApprenticeName { get ; private set ; }
+
+        [Name("Provider")]
+        public string Provider { get ; private set ; }
+
+        [Name("Apprenticeship training course")]
+        public string CourseName { get ; private set ; }
+
```
dashton82: Planned start date is missing but is included in #106

PR #104, opened 2020-02-27
MF-485 Added fix for max page number bug

robinoconnor:
Now when the page number exceeds the maximum value it will be reset to that value and hence always show the last results page


src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Controllers/ApprenticeControllerTests/ManageApprenticesTests.cs
```diff
@@ -53,6 +54,140 @@ public class ManageApprenticesTests
...
+
+            apprenticeshipMapper
+                .Setup(mapper => mapper.Map<IndexViewModel>(It.IsAny<IndexRequest>()))
+                .ReturnsAsync(expectedViewModel);
+
+            //Act
+            await controller.Index(request);
+
+            //Assert
+            cookieService.Verify(x => x.Update(CookieNames.ManageApprentices, request, It.IsAny<int>()));
```
dashton82: Should we also verify that the Get isn't called as well?

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/IndexRequest.cs
```diff
@@ -19,5 +19,6 @@ public int PageNumber
         public string SelectedCourse { get; set; }
         public DateTime? SelectedEndDate { get; set; }
         public ApprenticeshipStatus? SelectedStatus { get; set; }
+        public bool FromSearch { get; set; }
```
dashton82: shouldn't this be nullable?

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
```diff
@@ -1,45 +1,65 @@
...
+            _cookieStorage = cookieStorage;
+        }
+
+        [Route("", Name = RouteNames.ApprenticesIndex)]
+        [DasAuthorize(EmployerFeature.ManageApprenticesV2)]
+        public async Task<IActionResult> Index(IndexRequest request)
+        {
+            IndexRequest savedRequest = null;
+
+            if (request.FromSearch)
```
dashton82: I would have thought that this would be nullable

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/IndexRequest.cs
robinoconnor: We only need to know that its true to kick off the cookie search so making it null isn't really needed as it'll default to false

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
robinoconnor: No need as we just need to check for whether its true.

src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Controllers/ApprenticeControllerTests/ManageApprenticesTests.cs
robinoconnor: yep makes sense and no harm in checking it so its in.

PR #106, opened 2020-03-02
MF-497 download bug

dashton82:


src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
```diff
@@ -59,7 +60,28 @@ public async Task<IActionResult> Download(DownloadRequest request)
...
+                        {
+                            moreData = false;
+                        }
+
+                        downloadViewModel.Request.PageNumber += 1;
+
+                        stream2.CopyTo(outputStream);
+                    }
+
+                    downloadViewModel.Dispose();
```
scottywakefield: does this dispose before the next loop is executed? if so how do you use `downloadViewModel` again in the next loop?

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticeshipDetailsCsvModel.cs
```diff
@@ -17,6 +17,9 @@ public class ApprenticeshipDetailsCsvModel
         [Name("Apprenticeship training course")]
         public string CourseName { get ; private set ; }
 
+        [Name("Planned start date")]
+        public string PlannedStartDate { get; private set; }
```
scottywakefield: I'm guessing this is now wanted to be shown for employers? Is this only for download though as there's no cshtml change for it.

PR #114, opened 2020-04-01
Cv-684 levy status - NoneLevy companies not allowed to select Framework Courses

Pauljgraham:


src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Mappers/Cohort/ApprenticeViewModelMapperTests.cs
```diff
@@ -115,11 +120,21 @@ public void AutoCreatedReservationIsMappedCorrectly()
         }
 
         [Test]
-        public async Task TransferFundedCohortsAllowStandardCoursesOnly()
+        public async Task TransferFundedCohortsAllowStandardCoursesOnlyWhenEmployerIsLevy()
```
chrisfoster76, line 123: Is this test name change necessary? It seems to confuse things - only standards are allowed for transfers, irrespective of levy status. The "WhenEmployerIsLevy" bit in the name would imply a new setup step/precondition, but there is none.

chrisfoster76: Oh, I see. It's because of the new test below. Hmm. What if we combined them into a single test, accepting params for levy status, transfer or not, and whether frameworks expected? Might make the combinations clearer? Just a suggestion.

Pauljgraham: Thought about that, and it looked easier to read the test into 2 separate tests, otherwise the  I would have something like

AllowStandardCourseOnly(string transferSenderId, ApprenticeshipEmployerType levyStatus, bool hasFrameworks)

and it's slightly harder to understand.

PR #122, opened 2020-05-22
Cv 735 funding band message

hkf-tech:


src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/Cohort/DetailsViewModelMapper.cs
```diff
@@ -21,6 +21,7 @@ public class DetailsViewModelMapper : IMapper<DetailsRequest, DetailsViewModel>
         private readonly IEncodingService _encodingService;
         private readonly ITrainingProgrammeApiClient _trainingProgrammeApiClient;
         private readonly IAccountApiClient _accountsApiClient;
+        private bool IsLinkedToChangeOfPartyRequest;
```
chrisfoster76, line 24: It feels a little odd to have a property of the mapper track this. Is it just to avoid passing it to a method or two? Strictly speaking, is it really needed at all? The presence of an OriginalStartDate implies that this property be true anyway.

hkf-tech: Agree it is a bit odd, as we normally don't do it another mapper. i did it just to avoid multiple hops in methods to pass this value. 
But as you said not null OriginalStartDate implies it is LinkedToChangePartyRequest. I will remove IsLinkedToChangeOfPartyRequest and make it simpler.

PR #125, opened 2020-06-17
CV-771 analytics org name Employer

ChughAkshay:


src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Cohort/ApprenticeRequest.cs
```diff
@@ -2,10 +2,9 @@
 
 namespace SFA.DAS.EmployerCommitmentsV2.Web.Models.Cohort
 {
-    public class ApprenticeRequest : AssignRequest, IAuthorizationContextModel
+    public class ApprenticeRequest : AssignRequest
```
Pauljgraham, line 5: Curious why IAuthorizaruibContextModel was removed?  But I see it;s been added to the base class IndexRequest

src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Mappers/Cohort/WhenMappingSelectProviderRequestToViewModel.cs
```diff
@@ -1,86 +1,101 @@
...
+        [SetUp]
+        public void Setup()
         {
-            var result = await mapper.Map(request);
+            var autoFixture = new Fixture();
+            _request = autoFixture.Freeze<SelectProviderRequest>();
+            _commitmentsApiClientResponse = autoFixture.Freeze<AccountLegalEntityResponse>();
 
-            Assert.AreEqual(request.ReservationId, result.ReservationId);
+            _commitmentsApiClientMock = autoFixture.Freeze<Mock<ICommitmentsApiClient>>();
```
Pauljgraham, line 29: just curious why we can't just do 

_commitmentsApiClientMock = new Mock ICommitmentsApiClient ()

All the tests work, but would it work without Freeze? Just curious

src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Mappers/Cohort/WhenMappingSelectProviderRequestToViewModel.cs
```diff
@@ -1,86 +1,101 @@
...
+            _commitmentsApiClientMock
+                .Setup(x => x.GetAccountLegalEntity(_request.AccountLegalEntityId, CancellationToken.None))
+                .ReturnsAsync(_commitmentsApiClientResponse);
+
+            _mapper = new SelectProviderViewModelMapper(_commitmentsApiClientMock.Object);
         }
 
         [Test, MoqAutoData]
-        public async Task ThenMapsAccountHashedId(
```
Pauljgraham, line 25: If we are removing the injected autofixture variables then we should be able to remove the MoqAutoData attribute

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Cohort/ApprenticeRequest.cs
ChughAkshay: All the requests inherit from one another and I needed to get the AccountLegalEntityId to be populated in one of the base classes so I just moved the IAuthorizationContextModel to the top of the hierarchy, If it works like I think it works.

src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Mappers/Cohort/WhenMappingSelectProviderRequestToViewModel.cs
ChughAkshay: amended

src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Mappers/Cohort/WhenMappingSelectProviderRequestToViewModel.cs
ChughAkshay: amended

PR #137, opened 2020-08-05
CON-2210-add dataprotection keys to redis

shomavg:


src/SFA.DAS.EmployerCommitmentsV2.Web/Startup/AspNetStartup.cs
```diff
@@ -20,6 +24,10 @@ public AspNetStartup(IConfiguration configuration)
 
         public void ConfigureServices(IServiceCollection services)
         {
+            var connectionStrings = new ConnectionStrings();
```
chrisfoster76, line 27: I'm not sure this works; you need to `bind` the config, or, less elegantly, get it  through some means:
```
var redisConnectionString = _configuration.GetSection("SFA.DAS.EmployerCommitmentsV2").GetConnectionString("SharedRedis");
```

src/SFA.DAS.EmployerCommitmentsV2.Web/Startup/AspNetStartup.cs
```diff
@@ -20,6 +24,10 @@ public AspNetStartup(IConfiguration configuration)
...
+            var redis = ConnectionMultiplexer
```
chrisfoster76, line 28: Unless you have redis running at the configured location, you'll want to bypass this locally.

src/SFA.DAS.EmployerCommitmentsV2.Web/Startup/AspNetStartup.cs
```diff
@@ -28,7 +36,10 @@ public void ConfigureServices(IServiceCollection services)
                 .AddDasMvc()
                 .AddUrlHelper()
                 .AddEmployerUrlHelper()
-                .AddMemoryCache();
+                .AddMemoryCache()
+                .AddDataProtection()
+                .SetApplicationName("das-employercommitments-v2-web")
+                .PersistKeysToStackExchangeRedis(redis, "DataProtection-Keys");
```
chrisfoster76, line 42: As above, don't do this if running locally (check `_hostingEnvironment.IsDevelopment()` or similar)

src/SFA.DAS.EmployerCommitmentsV2.Web/Startup/AspNetStartup.cs
```diff
@@ -28,7 +31,8 @@ public void ConfigureServices(IServiceCollection services)
                 .AddDasMvc()
                 .AddUrlHelper()
                 .AddEmployerUrlHelper()
-                .AddMemoryCache();
+                .AddMemoryCache()
+                .SaveKeysToStackExchangeRedis(_configuration, Environment.IsDevelopment());
```
chrisfoster76, line 35: I would conditionally call this extension method, because the second parameter is basically a way of telling it not to do anything, which is not very readable, even if it adds a couple of lines here.

src/SFA.DAS.EmployerCommitmentsV2.Web/Startup/PersistDataStorageStartup.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Microsoft.AspNetCore.DataProtection;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using SFA.DAS.EmployerCommitmentsV2.Configuration;
+using SFA.DAS.EmployerCommitmentsV2.Web.Configurations;
+using StackExchange.Redis;
+
+namespace SFA.DAS.EmployerCommitmentsV2.Web.Startup
+{
+    public static class PersistDataStorageStartup
```
chrisfoster76, line 10: I'd recommend calling this DataProtectionStartupExtensions or something similar.

src/SFA.DAS.EmployerCommitmentsV2.Web/Startup/DataProtectionStartupExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using SFA.DAS.EmployerCommitmentsV2.Configuration;
+using SFA.DAS.EmployerCommitmentsV2.Web.Configurations;
+using StackExchange.Redis;
+
+namespace SFA.DAS.EmployerCommitmentsV2.Web.Startup
+{
+    public static class DataProtectionStartupExtensions
+    {
+        public static IServiceCollection SaveKeysToStackExchangeRedis(this IServiceCollection services, IConfiguration configuration, IHostingEnvironment environment)
```
chrisfoster76, line 13: Maybe this could be called `AddDataProtection`, as that's the first extension method that is called and fundamentally what is happening here.

PR #154, opened 2020-11-20
CON-2718 Adds validation to the input so user cannot select their cur…

ben1stone:
…rent provider


ben1stone:
I have also made changes to the API to stop the request if the new id is the same as the current Id

hkf-tech (approved):

PR #166, opened 2021-01-25
Update COP confirmation page

kola-samuel:


hkf-tech (approved):

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Apprentice/ChangeProviderInform.cshtml
```diff
@@ -14,13 +15,15 @@
     </h1>
     <div class="govuk-!-width-two-thirds das-panel__body">
 
-        @if (Model.ApprenticeshipStatus != ApprenticeshipStatus.Stopped)
+        @if (Model.ApprenticeshipStatus != ApprenticeshipStatus.WaitingToStart ||
+            Model.ApprenticeshipStatus != ApprenticeshipStatus.Live ||
+            Model.ApprenticeshipStatus != ApprenticeshipStatus.Paused)
```
ben1stone, line 20: This will always be true even if status is stopped, it should be == || ==

ben1stone (changes requested):

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Apprentice/ChangeProviderInform.cshtml
kola-samuel: good spot

ben1stone (approved):

PR #175, opened 2021-02-08
CON-3081 Install new package and add validation

ben1stone:


src/SFA.DAS.EmployerCommitmentsV2.Web/Validators/WhatIsTheNewStartDateViewModelValidator.cs
```diff
@@ -32,6 +38,11 @@ public WhatIsTheNewStartDateViewModelValidator()
                 .Must((r, newStartDate) => newStartDate.IsBeforeMonthYearOfDateTime(r.NewEndDate.Value))
                 .WithMessage(r => $"The start date must be before {r.NewEndDate:MMMM yyyy}")
                 .When(r => r.NewEndDate.HasValue);
+
+            RuleFor(r => r.NewStartDate)
```
hkf-tech, line 42: As mentioned on the das-commitment ticket, can we please change the code to something like , 

startDate <= _academicYear.CurrentAcademicYearEndDate.AddYears(1)

as the academicYear already has a CurrentAcademicYearEndDate property.
your code is correct, but to keep it aligned on what we have on V1 and to make it a bit easier to read with the user story i.e. ( I have entered a Month/ Year that is LATER THAN the end of current academic year + 1 Year (i.e., July ))

hkf-tech (changes requested):

src/SFA.DAS.EmployerCommitmentsV2.Web/Validators/WhatIsTheNewStartDateViewModelValidator.cs
ben1stone: Updated

hkf-tech (approved):

PR #181, opened 2021-02-24
Con 3172 flash message migration

VasanthaKasirajan3008:


src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
```diff
@@ -439,11 +442,13 @@ public async Task<IActionResult> PauseApprenticeship(PauseRequestViewModel viewM
             if (viewModel.PauseConfirmed.HasValue && viewModel.PauseConfirmed.Value)
             {
                 var pauseRequest = new PauseApprenticeshipRequest { ApprenticeshipId = viewModel.ApprenticeshipId };
-
+                
                 await _commitmentsApiClient.PauseApprenticeship(pauseRequest, CancellationToken.None);
+                
+                TempData.AddFlashMessage(ApprenticePausedMessage, ITempDataDictionaryExtensions.FlashMessageLevel.Info);
```
hkf-tech, line 448: Can you please add a flash message for stopped journey as well, as after stopping it may to go to other page instead of details.

hkf-tech (changes requested):

hkf-tech (approved):

PR #184, opened 2021-02-25
CON-3003 enhance co p confirmation when stopped in COP

kola-samuel:


cofaulco (commented):
All looks ok, my only question would be whether the StoppedDuringCoP property needs to be nullable

cofaulco (approved):

PR #200, opened 2021-04-29
CON-3334 Display change of provider history

ChrisJWoodcock:


src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticeshipDetailsRequestViewModel.cs
```diff
@@ -34,21 +35,17 @@ public class ApprenticeshipDetailsRequestViewModel : IAuthorizationContextModel
...
         public string HashedNewApprenticeshipId { get; set; }
-        public bool IsContinuation { get; set; }
-        public string HashedPreviousApprenticeshipId { get; set; }
         public bool ShowChangeTrainingProviderLink => ((ApprenticeshipStatus == ApprenticeshipStatus.Stopped ||
                                                        ApprenticeshipStatus == ApprenticeshipStatus.Paused ||
                                                        ApprenticeshipStatus == ApprenticeshipStatus.Live ||
                                                        ApprenticeshipStatus == ApprenticeshipStatus.WaitingToStart) &&
                                                        !HasPendingChangeOfProviderRequest &&
                                                        !HasPendingChangeOfEmployerRequest &&
-                                                       !(HasApprovedChangeOfEmployerRequest && !IsContinuation) &&
```
cofaulco, line 45: has this condition been correctly remove, meaning it was being applied incorrectly previously?

src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/Apprentice/ApprenticeshipDetailsRequestToViewModelMapper.cs
```diff
@@ -85,26 +86,35 @@ public async Task<ApprenticeshipDetailsRequestViewModel> Map(ApprenticeshipDetai
                     EmployerReference = apprenticeship.EmployerReference,
                     CohortReference = _encodingService.Encode(apprenticeship.CohortId, EncodingType.CohortReference),
                     EnableEdit = enableEdit,
-                    CanEditStatus = (apprenticeship.Status == ApprenticeshipStatus.Live || 
-                                     apprenticeship.Status == ApprenticeshipStatus.WaitingToStart || 
-                                     apprenticeship.Status == ApprenticeshipStatus.Paused),                    
+                    CanEditStatus = (apprenticeship.Status == ApprenticeshipStatus.Live ||
```
cofaulco, line 89: ApprenticeshipStatus is set a little further up this mapper, could this logic not just be a property getter? I know this is not something we have changed here. The whitespace changes just highlighted it to me

cofaulco (commented):
couple of comments

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticeshipDetailsRequestViewModel.cs
ChrisJWoodcock: Yes - the next line is using HashedNewApprenticeshipId which is now correctly obtained for both ChangeOfEmployer and ChangeOfProvider - this was a bug identified in the previous migration work and an AC was added to this ticket to fix it as the testing in in the same area

src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/Apprentice/ApprenticeshipDetailsRequestToViewModelMapper.cs
ChrisJWoodcock: Done

cofaulco (approved):

PR #215, opened 2021-06-04
Con 3346 choose organisation

VasanthaKasirajan3008:


src/SFA.DAS.EmployerCommitmentsV2.Web/Extensions/AgreementExtension.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Linq;
+using SFA.DAS.EAS.Account.Api.Types;
+using SFA.DAS.EmployerCommitmentsV2.Web.Models.Cohort;
+
+namespace SFA.DAS.EmployerCommitmentsV2.Web.Extensions
+{
+    public static class AgreementExtension
+    {   
+        public static bool HasSignedAgreement(this LegalEntity legalEntity, bool isTransfer)
```
cofaulco, line 9: could this name be a bit more obvious as to what it is doing? `HasSignedMinimumRequiredAgreementVersion` or something like that

src/SFA.DAS.EmployerCommitmentsV2.Web/Services/EmployerAccountsService.cs
```diff
@@ -0,0 +1,48 @@
...
+        }        
+
+        public async Task<List<LegalEntity>> GetLegalEntitiesForAccount(string accountId)
+        {
+            var listOfEntities = await _accountsApiClient.GetLegalEntitiesConnectedToAccount(accountId);
+
+            var list = new List<LegalEntity>();
+            if (listOfEntities.Count == 0) return list;
+
+            list.AddRange(listOfEntities.Select(async entity => await _accountsApiClient.GetLegalEntity(accountId, Convert.ToInt64(entity.Id))).Select(legalEntityViewModel => new LegalEntity
```
cofaulco, line 27: could make this parallel to improve the performance although would require some though as some accounts have >900 legal entities. Might be worth raising as a tech debt ticket.

cofaulco (commented):
a couple of minor points

src/SFA.DAS.EmployerCommitmentsV2.Web/Extensions/AgreementExtension.cs
VasanthaKasirajan3008: done

src/SFA.DAS.EmployerCommitmentsV2.Web/Services/EmployerAccountsService.cs
VasanthaKasirajan3008: tech debt raised : https://skillsfundingagency.atlassian.net/browse/CON-3606

cofaulco (approved):

PR #221, opened 2021-06-18
CON-3521 EAS Add Alert Filter

kola-samuel:


src/SFA.DAS.EmployerCommitmentsV2/SFA.DAS.EmployerCommitmentsV2.csproj
```diff
@@ -11,9 +11,9 @@
     <PackageReference Include="SFA.DAS.Authorization.CommitmentPermissions" Version="6.0.59" />
     <PackageReference Include="SFA.DAS.Authorization.EmployerFeatures" Version="6.0.59" />
     <PackageReference Include="SFA.DAS.Authorization.EmployerUserRoles" Version="6.0.59" />
-    <PackageReference Include="SFA.DAS.CommitmentsV2.Api.Client" Version="7.27.0" />
```
cofaulco, line 14: these versions are much lower than 7.29.34 in all the other project files?

cofaulco (changes requested):
1 change requested

src/SFA.DAS.EmployerCommitmentsV2/SFA.DAS.EmployerCommitmentsV2.csproj
kola-samuel: fixed

cofaulco (approved):

PR #225, opened 2021-06-24
Cs 632 add apprentice confirmation to display

Chazzysb:


src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Mappers/Apprentice/ApprenticeshipDetailsRequestToViewModelMapperTests.cs
```diff
@@ -90,7 +91,7 @@ public void SetUp()
             _mockEncodingService.Setup(t => t.Decode(It.IsAny<string>(), It.IsAny<EncodingType>()))
                 .Returns((string value, EncodingType encodingType) => long.Parse(Regex.Replace(value, "[A-Za-z ]", "")));
 
-            _mapper = new ApprenticeshipDetailsRequestToViewModelMapper(_mockCommitmentsApiClient.Object, _mockEncodingService.Object, Mock.Of<ILogger<ApprenticeshipDetailsRequestToViewModelMapper>>());
+            _mapper = new ApprenticeshipDetailsRequestToViewModelMapper(_mockCommitmentsApiClient.Object, _mockEncodingService.Object, Mock.Of<ILogger<ApprenticeshipDetailsRequestToViewModelMapper>>(), Mock.Of<IAuthorizationService>());
```
Pauljgraham, line 94: Should we add a test here to ensure ensure we map  ShowApprenticeConfirmationColumn and ConfirmationStatus correctly

Pauljgraham (changes requested):
Minor comment about adding a couple of tests to the mapper functionality

Pauljgraham (commented):
Couple of points.

1. Instead of creating a brand new test class to test these properties are mapping correctly, would it not have been easier to have simply have extended ApprenticeshipDetailsRequestToViewModelMapperTests.cs
to include a test that checked the property as mapped. 

2. In reality the responsibility of the mapper is to map the value from the source to the result, so we don't really care about the explicit value, we just care that it mapped teh source value correctly.

3. This doesn't actually have a test to cover the IAuthorizationService call to IsAuthorizedAsync

Pauljgraham (approved):

PR #233, opened 2021-07-14
CON-3544 Update information flash messages in EAS

Howesy77:


src/SFA.DAS.EmployerCommitmentsV2.Web/Extensions/ITempDataDictionaryExtensions.cs
```diff
@@ -19,6 +20,14 @@ public enum FlashMessageLevel
         public static void AddFlashMessage(this ITempDataDictionary tempData, string message, FlashMessageLevel level)
         {
             tempData[FlashMessageTempDataKey] = message;
```
cofaulco, line 22: could we make this `title` and `message` / `body`

cofaulco (commented):
1 comment

cofaulco (approved):

PR #237, opened 2021-07-16
CON-2777_Use_of_has_continuation

ChrisJWoodcock:


src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticeshipDetailsRequestViewModel.cs
```diff
@@ -35,17 +35,12 @@ public class ApprenticeshipDetailsRequestViewModel : IAuthorizationContextModel
...
                                                        ApprenticeshipStatus == ApprenticeshipStatus.Paused ||
                                                        ApprenticeshipStatus == ApprenticeshipStatus.Live ||
                                                        ApprenticeshipStatus == ApprenticeshipStatus.WaitingToStart) &&
-                                                       !HasPendingChangeOfProviderRequest &&
-                                                       !HasPendingChangeOfEmployerRequest &&
-                                                       string.IsNullOrEmpty(HashedNewApprenticeshipId));
-
-        public bool HasPendingChangeOfEmployerRequest { get; set; }
-        public Party? PendingChangeOfEmployerRequestWithParty { get; set; }
+                                                       !HasContinuation);
```
cofaulco, line 43: I think we need to check for the presence of a pending change of party request here too. Because HasContinuation will only be true if there is an approved Change of party.

cofaulco (commented):
1 comment

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticeshipDetailsRequestViewModel.cs
ChrisJWoodcock: HasContinuation is true for any row in [Apprenticeship] if there is another row for which the ContinuationOfId contains the Id regardless of whether the row IsApproved = 0 or 1; so HasContinuation is true when there is a Pending or Approved Change of Party. If a Change of Party is returned to it's originator and they delete it then the row in [Apprenticeship] is removed, so HasContinuation is then false.

cofaulco: 👍

cofaulco (approved):

PR #241, opened 2021-07-23
SV-991 transform training course name employer

ben1stone:


ben1stone:
Closed as this was completed with a new approach on das-commitments PR https://github.com/SkillsFundingAgency/das-commitments/pull/895

PR #244, opened 2021-08-09
Con 3734 uln overlap

hkf-tech:


src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Cohort/DetailsViewModel.cs
```diff
@@ -51,6 +51,17 @@ public string SendBackToProviderOptionMessage
                 return "No, request changes from training provider";
             }
         }
+
+        public bool HasOverlappingUln
+        {
+            get
+            {
+                return Courses != null
+                    && Courses.Count > 0
```
cofaulco, line 60: no need for the `count>0` line here

cofaulco (commented):
1 comment

cofaulco (approved):

PR #245, opened 2021-08-10
SV-425 select standard option page

ben1stone:


src/SFA.DAS.EmployerCommitmentsV2.Web/Views/DraftApprenticeship/SelectOption.cshtml
```diff
@@ -0,0 +1,75 @@
...
+            </div>
+            <div class="govuk-summary-list__row">
+                <dt class="govuk-summary-list__key"> Version </dt>
+                <dd class="govuk-summary-list__value"> @Model.Version </dd>
+            </div>
+        </dl>
+
+        <form id="selectOption" method="post">
+            <div class="govuk-form-group @Html.AddClassIfPropertyInError(x => x.CourseOption, "govuk-form-group--error")">
+                @Html.AntiForgeryToken()
```
dashton82, line 33: The Antiforgery is automatically added by the filter and isn't required here

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/DraftApprenticeship/SelectOption.cshtml
```diff
@@ -0,0 +1,75 @@
...
+                @Html.HiddenFor(x => x.DraftApprenticeshipId)
+                @Html.HiddenFor(x => x.FirstName)
+                @Html.HiddenFor(x => x.LastName)
+                @Html.HiddenFor(x => x.Email)
+                @Html.HiddenFor(x => x.Uln)
+                @Html.HiddenFor(x => x.CohortId)
+                @Html.HiddenFor(x => x.CohortReference)
```
dashton82, line 40: instead of `@Html.HiddenFor` can we use `<input type="hidden" asp-for="CohortReference"/>`

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CohortController.cs
```diff
@@ -254,6 +258,20 @@ public async Task<IActionResult> Apprentice(ApprenticeViewModel model)
             var request = await _modelMapper.Map<CreateCohortRequest>(model);
             var newCohort = await _commitmentsApiClient.CreateCohort(request);
 
+            var cohortApprenticeships = await _commitmentsApiClient.GetDraftApprenticeships(newCohort.CohortId);
+
+            if (cohortApprenticeships.DraftApprenticeships.Count() == 1)
```
CodeWolf86, line 263: Can probably change this to `SingleOrDefault` which gets you 1 if there is only 1 , or null if not, would then simplify your code below.

```
if(draftApprenticeship?.CourseCode != null)
{
  // encode and redirect
}
```

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/DraftApprenticeship/SelectOption.cshtml
```diff
@@ -0,0 +1,91 @@
...
+                <dd class="govuk-summary-list__value"> @Model.Version </dd>
+            </div>
+        </dl>
+
+    </div>
+</div>
+
+<form id="selectOption" method="post">
+    <div class="govuk-form-group @Html.AddClassIfPropertyInError(x => x.CourseOption, "govuk-form-group--error")">
+        <input type="hidden" asp-for="DraftApprenticeshipId" />
```
CodeWolf86, line 38: Do we need to re-post everything from the previous page? if the draft apprenticeship gets saved in the previous add, can we just pull through the id here to retrieve it on the server side post?

CodeWolf86 (changes requested):

CodeWolf86 (approved):

PR #257, opened 2021-09-03
SV-1165 upgrade version via CoC employer

ben1stone:


src/SFA.DAS.EmployerCommitmentsV2.Web/Mappers/Apprentice/ApprenticeshipDetailsRequestToViewModelMapper.cs
```diff
@@ -141,6 +159,21 @@ private static PendingChanges GetPendingChanges(CommitmentsV2.Api.Types.Response
...
-        }      
+        }
+
+        private async Task<bool> HasNewerVersions(TrainingProgramme trainingProgramme)
+        {
+            if (trainingProgramme.ProgrammeType == ProgrammeType.Standard)
+            {
+                var versionsResponse = await _commitmentsApiClient.GetTrainingProgrammeVersions(trainingProgramme.CourseCode);
+
+                if (versionsResponse.TrainingProgrammeVersions.Where(v => decimal.Parse(v.Version) > decimal.Parse(trainingProgramme.Version)).Count() > 0)
```
CodeWolf86, line 170: This will eventually fail for the same reason we did all the tech debt work with making version a string. 1.10 < 1.2 for example even though 1.10 is an actual version that's greater.

It's why I introduced VersionMajor, VersionMinor as columns against the cache, it may be we can do something with the Training Lookup Response to include them or factor them in to make this easier. Alternatively a newer versions endpoint which just returns true or false keeping that logic API side.

CodeWolf86 (changes requested):
See bugs to investigate on this ticket, 
https://skillsfundingagency.atlassian.net/browse/SV-1165?focusedCommentId=175746
Look for the PR Bugs Found / To Investigate title

CodeWolf86 (approved):

PR #268, opened 2021-10-04
CON-3819 Update Back link on Apprentice requests page

kola-samuel:


src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Cohort/Review.cshtml
```diff
@@ -51,5 +51,5 @@
 
 @section Back
 {
-    <div class="das-js-back-link"></div>
+    @Html.ActionLink("Back", "Index", "Home", new { @Model.AccountHashedId}, new { @class = "govuk-back-link" })
```
cofaulco, line 54: Link text should be "Back to Apprentices"

cofaulco (changes requested):
content change required

cofaulco (approved):

PR #287, opened 2022-01-25
CS-1354-ResendInvitation-Employer

Chazzysb:
CS-1354-ResendInvitation-Employer

### DIFF
--- a/src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
+++ b/src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
@@ -15,6 +15,7 @@ using SFA.DAS.EmployerCommitmentsV2.Web.Cookies;
 using SFA.DAS.EmployerCommitmentsV2.Web.Extensions;
 using SFA.DAS.EmployerCommitmentsV2.Web.Models.Apprentice;
 using SFA.DAS.EmployerCommitmentsV2.Web.Models.Apprentice.Edit;
+using SFA.DAS.EmployerCommitmentsV2.Web.Models.Shared;
 using SFA.DAS.EmployerCommitmentsV2.Web.RouteValues;
 using SFA.DAS.EmployerCommitmentsV2.Web.Authentication;
 using EditEndDateRequest = SFA.DAS.EmployerCommitmentsV2.Web.Models.Apprentice.EditEndDateRequest;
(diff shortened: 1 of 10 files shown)


src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
```diff
@@ -858,5 +862,19 @@ public async Task<IActionResult> DataLockRequestRestart(DataLockRequestRestartRe
...
+        [HttpGet]
+        public async Task<IActionResult> ResendEmailInvitation([FromServices] IAuthenticationService authenticationService, ResendEmailInvitationRequest request)
+        {
+            await _commitmentsApiClient.ResendApprenticeshipInvitation(request.ApprenticeshipId, new SaveDataRequest { UserInfo = authenticationService.UserInfo }, CancellationToken.None);
+
+            return RedirectToAction("ApprenticeshipDetails", new {
+                AccountHashedId = request.AccountHashedId,
+                ApprenticeshipHashedId = request.ApprenticeshipHashedId,
+                ShowNotification = true,
+                NotificationTitle = "The invitation email has been resent."
```
Pauljgraham, line 876: hopefully these last 2 params will not be needed

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
```diff
@@ -858,5 +862,19 @@ public async Task<IActionResult> DataLockRequestRestart(DataLockRequestRestartRe
...
+            });
```
Pauljgraham, line 877: We should be able to put 

TempData.AddFlashMessage(notification.NotificationTitle, notification.NotificationBody, ITempDataDictionaryExtensions.FlashMessageLevel.Success);

here

Pauljgraham (commented):
I think we should be able to simplify the flash message. Please see below.

I'm not sure if we can actually test the new endpoint. I would have thought it would be possible

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
```diff
@@ -563,10 +564,13 @@ public async Task<IActionResult> ResumeApprenticeship(ResumeRequestViewModel vie
...
         [DasAuthorize(CommitmentOperation.AccessApprenticeship)]
         [Route("{apprenticeshipHashedId}/details", Name = RouteNames.ApprenticeDetail)]
-        public async Task<IActionResult> ApprenticeshipDetails(ApprenticeshipDetailsRequest request)
+        public async Task<IActionResult> ApprenticeshipDetails(ApprenticeshipDetailsRequest request, NotificationParameters notification = null)
         {
             var viewModel = await _modelMapper.Map<ApprenticeshipDetailsRequestViewModel>(request);
 
+            if (notification?.ShowNotification == true)
+                TempData.AddFlashMessage(notification.NotificationTitle, notification.NotificationBody, ITempDataDictionaryExtensions.FlashMessageLevel.Success);
+
```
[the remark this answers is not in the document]
Chazzysb: Moved the flash message call to the resend route

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
Chazzysb: removed

Pauljgraham (approved):

PR #302, opened 2022-03-01
PA-187 - Finalise enum names for Delivery Model

benagain:


src/SFA.DAS.EmployerCommitmentsV2.Web/Extensions/DeliveryModelExtension.cs
```diff
@@ -10,14 +10,14 @@ public static class DeliveryModelExtension
...
-                _ => "Normal"
+                DeliveryModel.PortableFlexiJob => "Portable flexi-job",
+                _ => "Regular"
             };
 
         public static string ToAbnormalDescription(this DeliveryModel deliveryModel) =>
             deliveryModel switch
             {
-                DeliveryModel.Flexible => "Flexi-job",
+                DeliveryModel.PortableFlexiJob => "Portable flexi-job",
```
Pauljgraham, line 20: Again can we change the exyension name to ToIrregual.....

Pauljgraham (approved):
Extension name suggestion

PR #305, opened 2022-03-29
Do not allow editing of approved flexi-job apprenticeships

benagain:


Pauljgraham (changes requested):
I think we need to stop the user from editing, the delivery model when the course is regular. Just need to stop the Delivery Model from being editted on the view

Pauljgraham (approved):

PR #307, opened 2022-04-13
PA-202-AddSelectCourseAndDeliveryModelPages

Chazzysb:
PA-202-AddSelectCourseAndDeliveryModelPages

### DIFF
--- a/src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CohortController.cs
+++ b/src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CohortController.cs
@@ -1,8 +1,4 @@
-﻿using System;
-using System.Net;
-using System.Threading;
-using System.Threading.Tasks;
-using Microsoft.AspNetCore.Mvc;
+﻿using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 using SFA.DAS.Authorization.CommitmentPermissions.Options;
 using SFA.DAS.Authorization.EmployerUserRoles.Options;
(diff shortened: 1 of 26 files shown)


src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CohortController.cs
```diff
@@ -256,23 +260,98 @@ private Origin DetermineOrigin(AssignViewModel source)
...
-            return View(model);
+            //if (_authorizationService.IsAuthorized(EmployerFeature.DeliveryModel))
+            //{
+                return RedirectToAction(nameof(SelectCourse), request);
+            //}
+            //else
+            //{
+                //return RedirectToAction(nameof(AddDraftApprenticeship), request);
+            //}
+        }
```
Pauljgraham, line 273: We might need the toggle on, but I'm not 100% sure

Pauljgraham (commented):
Looks very good, a couple of small comments. It's a good foundation for us to build on top of

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CohortController.cs
```diff
@@ -256,23 +260,98 @@ private Origin DetermineOrigin(AssignViewModel source)
...
+              var  model = await _modelMapper.Map<ApprenticeViewModel>(request);
+            //}
+            //else
+            //{
+            //    model.CourseCode = request.CourseCode;
+            //    model.DeliveryModel = request.DeliveryModel;
+            //}
+            return View("Apprentice", model);
+        }
+
```
Pauljgraham, line 343: The above methods are still missing tests

PR #308, opened 2022-04-21
Pa 220 add apprentice display new rows layout

Chazzysb:
Pa 220 add apprentice display new rows layout


src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/CohortController.cs
```diff
@@ -262,14 +262,14 @@ private Origin DetermineOrigin(AssignViewModel source)
...
-            //}
-            //else
-            //{
-                //return RedirectToAction(nameof(AddDraftApprenticeship), request);
-            //}
+            }
+            else
+            {
+                return RedirectToAction(nameof(AddDraftApprenticeship), request);
+            }
```
Pauljgraham, line 272: I can't see tests for this action

Pauljgraham (commented):
Looks good but we are missing tests for the controller actions. I would love to be able to test these the way we test ApprenticeCommitments, but we are stuck with an old fashioned way of working

PR #309, opened 2022-04-22
CON 4554 remove redundant features

hasanrq:


src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Controllers/DraftApprenticeshipControllerTests/AddDraftApprenticeshipTests.cs
```diff
@@ -128,8 +128,6 @@ public AddDraftApprenticeshipTestsFixture()
             AuthorizationService = new Mock<IAuthorizationService>();
```
cofaulco, line 128: this can be removed as the controller no longer requires the dependency

src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Controllers/DraftApprenticeshipControllerTests/AddDraftApprenticeshipTests.cs
```diff
@@ -128,8 +128,6 @@ public AddDraftApprenticeshipTestsFixture()
...
             EncodingService = new Mock<IEncodingService>();
 
-            AuthorizationService.Setup(x => x.IsAuthorized(EmployerFeature.EnhancedApproval)).Returns(false);
```
cofaulco, line 131: Also a using statement at the top of this class which can be removed

cofaulco (changes requested):
If you remove the `IAuthorizationService` dependency from the constructor of `DraftApprenticeshipController` then the unit test Mocks associated can be cleaned up

src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Controllers/DraftApprenticeshipControllerTests/AddDraftApprenticeshipTests.cs
hasanrq: this has been updated

src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Controllers/DraftApprenticeshipControllerTests/AddDraftApprenticeshipTests.cs
hasanrq: this has been removed

PR #311, opened 2022-04-27
Pa 234 add another apprentice

Chazzysb:
Pa 234 add another apprentice


Pauljgraham (commented):
Looking good, only the small bug (when going via options) and the required tests on the controller actions. You will find a lots of these present in PAS and you should be able to just lift most of the code over.

The only other issue, is adding the CohortId and AccountLegalEntityId to the URLs. They shouldn't be there (as they are not supposed to be exposed outsdide).

PR #322, opened 2022-06-23
PA-372 Added support to show DeliveryModel link if there are multiple

Pauljgraham:


src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Apprentice/EditApprenticeship.cshtml
```diff
@@ -13,8 +16,9 @@
     ViewBag.Section = "apprentices";
     ViewBag.PageId = "edit-apprentices-detail";
     ViewBag.GaData.Vpv = "/accounts/apprentices/manage/edit-apprentice-details";
+    var multipleDeliveryModels = await DeliveryModelService.HasMultipleDeliveryModels(Model.ProviderId, Model.CourseCode);
```
chrisfoster76, line 19: @Pauljgraham @Chazzysb You might want to consider not implementing this change this way. Why not add the required property to the viewmodel in the usual way? Injecting an api client wrapped into a service just to get a piece of information for the page seems like a hack.

chrisfoster76 (dismissed):
See comments on implementation

PR #325, opened 2022-07-15
FJA 66 - As per ticket only show fields if portable delivery model

alireid:


chrisfoster76 (approved):

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/DraftApprenticeship/View.cshtml
```diff
@@ -65,7 +66,7 @@
             {
```
chrisfoster76, line 66: The Delivery Model should be hidden if it's `Regular`. Most delivery models are Regular, and where this is the case most users won't even know what it means.

chrisfoster76 (changes requested):

src/SFA.DAS.EmployerCommitmentsV2.Web/Views/DraftApprenticeship/View.cshtml
alireid: Fixed in lastest commit

chrisfoster76 (approved):

PR #328, opened 2022-07-27
PA-432-RemoveIdsFromQuerystring

Chazzysb:
PA-432-RemoveIdsFromQuerystring

### DIFF
--- a/src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Cohort/AssignRequest.cs
+++ b/src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Cohort/AssignRequest.cs
@@ -1,6 +1,24 @@
-﻿namespace SFA.DAS.EmployerCommitmentsV2.Web.Models.Cohort
+﻿using Microsoft.AspNetCore.Mvc;
+using SFA.DAS.EmployerCommitmentsV2.Web.Extensions;
+
+namespace SFA.DAS.EmployerCommitmentsV2.Web.Models.Cohort
 {
-    public class AssignRequest : ConfirmProviderRequest
+    public class AssignRequest : BaseAssignRequest
     {
+        public long AccountLegalEntityId { get; set; }
+    }
+
+    public class BaseAssignRequest : IndexRequest
+    {
+        public long ProviderId { get; set; }
+        public string LegalEntityName { get; set; }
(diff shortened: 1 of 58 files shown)


src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Cohort/AssignRequest.cs
```diff
@@ -1,6 +1,24 @@
...
 {
-    public class AssignRequest : ConfirmProviderRequest
+    public class AssignRequest : BaseAssignRequest
     {
+        public long AccountLegalEntityId { get; set; }
+    }
+
+    public class BaseAssignRequest : IndexRequest
+    {
+        public long ProviderId { get; set; }
```
benagain, line 13: I think `ProviderId` should be hidden from the URL as well, shouldn't it?

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Cohort/ConfirmProviderRequest.cs
```diff
@@ -1,8 +1,24 @@
...
 {
-    public class ConfirmProviderRequest : SelectProviderRequest
+    public class ConfirmProviderRequest : BaseConfirmProviderRequest
+    {
+        public long AccountLegalEntityId { get; set; }
+    }
+
+    public class BaseConfirmProviderRequest : IndexRequest
     {
         public long ProviderId { get; set; }
```
benagain, line 13: Again, I think ProviderId should be hidden from the URL as well.

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Cohort/AssignRequest.cs
Chazzysb: provider id is entered by the user i the journey

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Cohort/ConfirmProviderRequest.cs
Chazzysb: provider id is entered by the user in the journey

Pauljgraham (approved):

PR #352, opened 2022-09-27
PA-408 add delivery model to csv export

Pauljgraham:


src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticeshipDetailsCsvModel.cs
```diff
@@ -57,7 +59,8 @@ public class ApprenticeshipDetailsCsvModel
                 TotalAgreedPrice = $"{model.TotalAgreedPrice.Value as object:n0}",
                 ApprenticeConfirmation = model.ConfirmationStatus.ToDisplayString(),
                 Status = model.ApprenticeshipStatus.GetDescription(),
-                Alerts = GenerateAlerts(model.Alerts)
+                Alerts = GenerateAlerts(model.Alerts),
+                DeliveryModel = ToDeliveryModelDescription(model.DeliveryModel)
```
benagain, line 63: You should include the trailing comma so that future diffs don't have the problem this one has of highlighting an irrelevant line.

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticeshipDetailsCsvModel.cs
```diff
@@ -76,5 +79,16 @@ private static string GenerateAlerts(IEnumerable<Alerts> alerts)
 
             return alertString;
         }
+
+        private static string ToDeliveryModelDescription(DeliveryModel deliveryModel)
+        {
+            return deliveryModel switch
+            {
+                CommitmentsV2.Types.DeliveryModel.FlexiJobAgency => "Flexi-job agency",
```
benagain, line 87: It would be better to use the existing conversion rather than duplicate it

https://github.com/SkillsFundingAgency/das-employercommitments-v2/blob/752861079271d030e7835f658c257b9f6ea100c7/src/SFA.DAS.EmployerCommitmentsV2.Web/Extensions/DeliveryModelExtension.cs#L10-L16

Pauljgraham: yes good point.

benagain (approved):

PR #358, opened 2022-10-12
Readme content change

ratheeshri1984:


README.md
```diff
@@ -13,5 +13,6 @@ Functionality within EmployerCommitmentsV2 requires Commitments Api v2 (https://
   * SFA.DAS.EmployerUrlHelper
   * SFA.DAS.Encoding
   * SFA.DAS.Reservations.Api.Client
+  * SFA.DAS.EmployerAccounts.Api.Client
```
cofaulco, line 16: Are you sure this is correct? This service should not require that config rowkey. It calls the api in question, however the details for local running of that api service i.e config required should be in the repo of that application.

cofaulco (changes requested):

README.md
cofaulco: OK I see this is required because of the das-authorization service.

PR #360, opened 2022-10-13
Map_111OLTD stopped Apprenticeship

kola-samuel:


src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
```diff
@@ -999,5 +999,48 @@ public async Task<IActionResult> ReconfirmHasNotStopChangesAsync(ReconfirmHasNot
...
+        {
+            var viewModel = await _modelMapper.Map<ConfirmWhenApprenticeshipStoppedViewModel>(request);
+            return View(viewModel);
+        }
+
+        [HttpPost]
+        [Route("{apprenticeshipHashedId}/details/confirmWhenApprenticeshipStopped")]
+        public async Task<IActionResult> ConfirmWhenApprenticeshipStopped(ConfirmWhenApprenticeshipStoppedViewModel viewModel)
+        {
+            if (viewModel.IsCorrectStopDate.HasValue)
```
cofaulco, line 1015: this check is handled by the Model validation in `ConfirmWhenApprenticeshipStoppedViewModelValidator`

cofaulco (changes requested):
1 minor comment

src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
kola-samuel: changed

cofaulco (approved):

PR #372, opened 2022-11-07
FLP-119 Navigate to the [Add/Edit] training details page from the [Ad…

ruslanss:
…d/Edit] personal details page when the keyboard ‘Enter’ button is used

### DIFF
--- a/src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Shared/_DraftApprenticeship.cshtml
+++ b/src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Shared/_DraftApprenticeship.cshtml
@@ -444,7 +444,6 @@
         validateDraftApprenticeship(submitForm);
     }
 
-
     function submitForm($form) {
         $form.off('submit');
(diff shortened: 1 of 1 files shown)


src/SFA.DAS.EmployerCommitmentsV2.Web/Views/Shared/_DraftApprenticeship.cshtml
```diff
@@ -460,4 +459,27 @@
 
         return indexed_array;
     }
+
+    
+    function onFormSubmitted(e) {
+        e.preventDefault();
+        if ($("#draftApprenticeshipSection1").hasClass("das-hide") === false)
+            continueJourney();
+        else return;
```
tom-gough, line 468: Does returning in this function end up calling the default form submit behaviour?

ruslanss: yup

tom-gough (approved):

PR #398, opened 2023-01-20
Fields not displaying on review page

alireid:
They must only show if the updates are using portable


Pauljgraham (approved):
Approved, but TBH I cannot see why the old code doesn't work! The logic looks the same to me

PR #409, opened 2023-02-20
MAP-217-Additional fields become editable on PAUSED record

ratheeshri1984:


src/SFA.DAS.EmployerCommitmentsV2.Web.UnitTests/Mappers/Apprentice/EditApprenticeshipRequestToViewModelMapperTests.cs
```diff
@@ -313,6 +312,36 @@ public async Task IsLockedForUpdate_Is_Mapped_With_ApprenticeshipStatus_And_IsNo
             Assert.AreEqual(expectedIsLockedForUpdated, viewModel.IsLockedForUpdate);
         }
 
+        [TestCase(ApprenticeshipStatus.Paused, true)]
+        public async Task IsLockedForUpdate_And_IsLive_Is_Mapped_With_ApprenticeshipStatus_And_IsNotWithInFundingPeriod_Condition(ApprenticeshipStatus status, bool expectedIsLockedForUpdated)
```
cofaulco, line 316: This test has been split out and renamed to a misleading name for no good reason? Should have just been a changed to the expected result in the above test case

cofaulco (changes requested):

cofaulco (approved):

PR #445, opened 2024-01-17
Add banner when price change rejected

MichaelYoung1981:


src/SFA.DAS.EmployerCommitmentsV2.Web/Controllers/ApprenticeController.cs
```diff
@@ -621,9 +621,10 @@ public async Task<IActionResult> ResumeApprenticeship(ResumeRequestViewModel vie
         [HttpGet]
         [DasAuthorize(CommitmentOperation.AccessApprenticeship)]
         [Route("{apprenticeshipHashedId}/details", Name = RouteNames.ApprenticeDetail)]
-        public async Task<IActionResult> ApprenticeshipDetails(ApprenticeshipDetailsRequest request)
+        public async Task<IActionResult> ApprenticeshipDetails(ApprenticeshipDetailsRequest request, bool showPriceChangeRejected)
```
tom-gough, line 624: Looks like the unit test for this method needs updating following adding this parameter

tom-gough (changes requested):

tom-gough (approved):

PR #446, opened 2024-01-23
Flp 572 generate bearer token

MikeV-TalentConsulting:


src/SFA.DAS.EmployerCommitmentsV2/Services/BearerTokenProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+			if (string.IsNullOrEmpty(signingKey))
+			{
+				throw new BearerTokenException("Signing key cannot be null or empty");
+			}
+
+			const int minimumKeySize = 128;
+			if (signingKey.Length * 8 < minimumKeySize)
+			{
+				// This checks the key is at least 128 bits long, otherwise the token will fail to be generated
+				throw new BearerTokenException("Signing key must exceed 128bits in length");
```
tom-gough, line 28: can we change exceed to "be at least" again?

PR #512, opened 2024-09-09
APPMAN-655 remove tempdata

ogrwill:


Pauljgraham (commented):
Managed to test it locally for cohorts and draft apprenticeships, but couldn't get our base code working for approved apprenticeships (nothing to do with your work). 

Looks good, just a few minor issues

Pauljgraham (commented):
Looks ok, (not an easy job to replace it) I left a few comments about deleting the cache after we read it as this mirrors TempData. but I think I'm wrong about this. As it will hinder the back button.

PR #641, opened 2026-06-09
Appman 2561 freeze payments

cofaulco:


src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticeshipDetailsRequestViewModel.cs
```diff
@@ -31,9 +31,11 @@ public class ApprenticeshipDetailsRequestViewModel : IAuthorizationContextModel
     public string ProviderName { get; set; }
     public PendingChanges PendingChanges { get; set; }
 
-    public bool CanEditStatus => (ApprenticeshipStatus == ApprenticeshipStatus.Live ||
-                                  ApprenticeshipStatus == ApprenticeshipStatus.WaitingToStart ||
-                                  ApprenticeshipStatus == ApprenticeshipStatus.Paused);
+    public bool CanEditStatus => false;
```
Pauljgraham: I don't think we can just turn set the CanEditStatus to false. Employers can still stop an apprenticeship as long as it wasn't stopped via the ILR withdrawn event), but you will need to change that journey to exclude the pause option

Pauljgraham (commented):
Really good, but the CanEditStatus can not yet be set to false

Pauljgraham (approved):

src/SFA.DAS.EmployerCommitmentsV2.Web/Models/Apprentice/ApprenticeshipDetailsRequestViewModel.cs
cofaulco: reverted

